PR #8, opened 2023-12-01
Comment out debug prints in non-test code

darosior:
Those could be converted to logs instead as they are still useful debugging info. Since there is no log enginge, simply comment them out like the rest of the debug prints in the codebase.


evanlinjin (approved):
ACK 247e3b6253b45bed4673602c4b10c7c84d36313b

Issue #6, opened 2023-11-30
Make `LowestFee` metric better


evanlinjin:
- https://github.com/bitcoindevkit/bdk/pull/1072#discussion_r1393680784
- https://github.com/bitcoindevkit/bdk/pull/1072#discussion_r1393687511

LLFourn:
I did this in #13. It is certainly "better" (as in not completely broken) now :)

PR #11, opened 2023-12-12
fix: add excess to changeless `LowestFee` score

jp1ac4:
This is to fix the part of #6 relating to https://github.com/bitcoindevkit/bdk/pull/1072#discussion_r1393687511.

For the `LowestFee` metric, it adds the excess to changeless solutions instead of those with change.


LLFourn:
I still think it's wrong after this. Why are we adding the lower bound thing to it. In the case of no drain the metric should just be:   `(cs.selected_value() - self.target.value) as f32` i.e. the fee. Then with drain it should be `(input_weight + base_weight + drain_weight) * feerate`. 

cc @evanlinjin

LLFourn:
Just noting that I've talked with @evanlinjin and I'm now in charge of fixing it.

LLFourn:
@evanlinjin can you close this PR and give me permission to manage this repo.

PR #14, opened 2023-12-21
Simplify change policy

LLFourn:
This PR makes a simplifying assumption: The decision about whether to add change or not can be reduced to a check whether the excess is over a certain threshold. The change policy is simply how to compute that threshold.

I originally didn't do this because I thought change policies could be more complicated than that -- and they probably could be! But @evanlinjin pointed out that you can easily run the coin selector multiple times with different policies and change output structures if you wanted to compare and contrast different approaches (rather than trying to embed this logic into the change policy itself).

I realized it is actually quite difficult to implement the metrics correctly without knowing the threshold that will case change to be added (and the weight of that change) up front. Knowing this we should be able to make the implementations much simpler.

### DIFF
--- a/src/coin_selector.rs
+++ b/src/coin_selector.rs
@@ -625,9 +625,9 @@ impl Candidate {
     }
 }
 
-/// A structure that represents the weight costs of a drain (a.k.a. change) output.
+/// Represents the weight costs of a drain (a.k.a. change) output.
 ///
-/// This structure can also represent multiple outputs.
+/// May also represent multiple outputs.
 #[derive(Default, Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub struct DrainWeights {
     /// The weight of including this drain output.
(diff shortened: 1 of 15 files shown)


evanlinjin (commented):
Concept ACK dac064173470808a289abbb06c5c1acf18939c3d

src/coin_selector.rs
```diff
@@ -419,6 +429,40 @@ impl<'a> CoinSelector<'a> {
...
+    /// If not change output should be added according to policy then it will return `None`.
+    pub fn drain_value(&self, target: Target, change_policy: ChangePolicy) -> Option<u64> {
+        let excess = self.excess(
+            target,
+            Drain {
+                weights: change_policy.drain_weights,
+                value: 0,
+            },
+        );
+        if excess > change_policy.min_value as i64 {
```
jp1ac4, line 444: I'm wondering if this should have been `>=` instead of `>` . The `min_value_and_waste` change policy was previously returning change for excess equal to the min value.

LLFourn: I think it should be `>=` indeed.

Issue #12, opened 2023-12-14
Allow to target suitable transaction fee for RBF


jp1ac4:
See https://github.com/bitcoindevkit/bdk/pull/1072#issuecomment-1819593099 and follow-up comments.

PR #19, opened 2024-01-24
Implement proper RBF API so you can satisfy rule 4

LLFourn:
Removes `min_fee` constraint in favour of proper rule 4 handling (min_fee was pretty useless).

See for rule 4:
https://github.com/bitcoin/bitcoin/blob/master/doc/policy/mempool-replacements.md#current-replace-by-fee-policy

LowestFee BnB metric updated to bound correctly for this.

A few other coin_selector API changes

Fixes: #12


tests/changeless.rs
```diff
@@ -14,7 +14,7 @@ fn test_wv(mut rng: impl RngCore) -> impl Iterator<Item = Candidate> {
             value,
             weight: rng.gen_range(0..100),
             input_count: rng.gen_range(1..2),
-            is_segwit: rng.gen_bool(0.5),
+            is_segwit: false,
```
LLFourn, line 17: Note we never want to choose this randomly in the proptests because we cannot accurately lower bound it.

evanlinjin (approved):
ACK 0f7cc313b792c4b4a9b58ece8396f02f71191d77

This is excellent.

PR #21, opened 2024-01-25
No more base weight

### LINKED ISSUE #1: Simplify `DrainWeights`/`Target` API
Original discussion: https://github.com/bitcoindevkit/bdk/pull/1072#issuecomment-1802704053

@jp1ac4 @darosior I'm working on some breaking changes to the API.

After a discussion with @LLFourn, we've decided to include `DrainWeights` in `Target` and keep `Target` in the `CoinSelector`. This simplifies the API and gives us an easy pathway to take into account the output count varint weight changes when including the drain output.

How this simplifies the API:

* The target & drain weights are included in the `CoinSelector`, so they don't need to be passed in as method inputs.
* Change policy can be simplified to just: `Fn(&CoinSelector) -> bool`.

How we can take into account the output varint weight changes when including drain:

```rust
pub struct Target {
    /*other fields*/

    pub base_weight: u32,
    pub drain_weights: DrainWeights,
}

impl Target {
    pub fn new(
        feerate: FeeRate,
        recipient_outputs: impl IntoIterator<Item = (u32 /*weight*/, u64 /*value*/)>
        drain_outputs: impl IntoIterator<Item = DrainWeights>,
    ) -> Self {
        todo!()
    }
}
```

This calculates the `base_weight` and the `drain_weight` (which includes output varint changes) that are both included in `Target`.

LLFourn:
Fixes #1

On top of #19 

- CoinSelector no longer tracks anything but input weight
- Previously the value of the target outputs was in `Target` but the
  weights were accounted for in CoinSelector. Now they're in all in
  target.
- This allows us to actually figure out how many outputs there are and
  therefore the actual weight of the transaction accounting for the varint for the number of outputs.

This wasn't what the issue had in mind but it was easier to take the `base_weight` out of `CoinSelector` and put it in `Target` rather than put `Target` in `CoinSelector`. Getting rid of `base_weight` is a more crucial change than expected because rust bitcoin changed what `Transaction::weight` returns for empty output transactions recently so using it to determine `base_weight` will get different answers between versions (this breaks our weight tests but this PR will fix it I think if we uprade dev deps). We only need to know the total weight of the outputs and how many there are now to get the right answers for weight.


murchandamus:
Yeah, the waste metric is only used as a tiebreaker rather than something to optimize for because it would happily consolidate all your UTXOs at low feerates. What did you end up using instead?

LLFourn:
Hey @murchandamus! We've been keeping this code base low profile because it's still in science experiment stage. FWIW we gathered a long time ago that the waste metric was not something that you should use to actually optimize for -- it was just a good thing one to use to start the experiment.  I'm interested in your work here: https://delvingbitcoin.org/t/gutterguard-and-coingrinder-simulation-results/279. It would be cool to try and replicate it 

We're just using what we call `LowestFee` metric: Tries to minimize the fee cost of the selection while taking into account the cost of spending the change (using the "long term feerate").

murchandamus:
Ah that sounds similar to what I’m trying to do with [CoinGrinder](https://github.com/bitcoini/bitcoin/pull/27877), where I am trying to always find the input set with the lowest input weight. The PR should hopefully be very readable: I organized the pullrequest to start with a simple implementation of the algorithm and then each improvement or optimization a separate commit. I still have a little hope that it might make the release, even if I did miss the really big feerates. In combination with BnB, and the waste metric, I guess I’d get something similar as the outcome of your `LowestFee` metric.

evanlinjin (approved):
ACK ab38cb00c11cb30b531b07aa37765647876f5579

I saw some obvious mistakes so I pushed fixes as commits. Everything else looks good to me. The `LowestFee` bound function looks correct.

evanlinjin (approved):
ACK c6503060454e4cf8e04990c89c9e30177af77d17

PR #25, opened 2024-04-04
docs: fix varint comment for `DrainWeights`

jp1ac4:
Following #21, the `output_weight` in `DrainWeights` must not take into account a change in varint output count as this is now handled elsewhere.


LLFourn (approved):
Ahhh nice catch.

PR #28, opened 2024-09-07
feat!: use `u64` for weights instead of `u32`

evanlinjin:
Also change tests and examples to use `v0.32` of the `bitcoin` crate.


LLFourn:
ACK 72761ae0ea274a9f213a5fe4b44ff0d41c8fc4fb

PR #29, opened 2024-10-29
Calculate transaction fee by applying fee rate to rounded-up vbytes

### LINKED ISSUE #26: Option to calculate fee based on rounded-up vbytes
When targeting a fee rate in sat/vb, could there be the option to calculate the final fee based on the rounded-up vbytes, i.e.
`vb = (wu / 4).ceil()`.

Currently, when using the rounded-up vbytes to calculate the fee rate after using coin selection, we can end up with a transaction paying a slightly lower value than our target.

I think it might be enough to modify the method `CoinSelector::implied_fee_from_feerate` so that it can optionally round up the vbytes. It should be sufficient to do the rounding up at the end once the selection is complete, and not for intermediate calculations, e.g. when determining the cost of adding a single input, as we don't want to round up once for each input.

Perhaps this option could be set as part of `TargetFee`.

If you think this would be fine to add, I'd be happy to work on it. Thanks.

jp1ac4:
This is an attempt to resolve #26.

When calculating the implied fee of a transaction, it applies the fee rate in sat/vb to the transaction's rounded-up vbytes, including for RBF rule 4 considerations.

### DIFF
--- a/tests/rbf.rs
+++ b/tests/rbf.rs
@@ -5,7 +5,7 @@ fn run_bitcoin_core_rbf_tests() {
     // see rbf_tests.cpp
     //
     // https://github.com/bitcoin/bitcoin/blob/e69796c79c0aa202087a13ba62d9fbcc1c8754d4/src/test/rbf_tests.cpp#L151
-    const CENT: u64 = 100_000; // no clue why this would be called CENT 😕
+    const CENT: u64 = 1_000_000;
     let low_fee = CENT / 100;
     let _normal_fee = CENT / 10;
     let high_fee = CENT;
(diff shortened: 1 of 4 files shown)


tests/rbf.rs
```diff
@@ -5,65 +5,78 @@ fn run_bitcoin_core_rbf_tests() {
     // see rbf_tests.cpp
     //
     // https://github.com/bitcoin/bitcoin/blob/e69796c79c0aa202087a13ba62d9fbcc1c8754d4/src/test/rbf_tests.cpp#L151
-    const CENT: u64 = 100_000; // no clue why this would be called CENT 😕
+    const CENT: u64 = 1_000_000;
```
LLFourn, line 8: Thanks for catching this. I suppose I understand what CENT means now :facepalm:

LLFourn (approved):
ACK 4eae611676c41628080eeb5d1a48a5abd841c72c

PR #30, opened 2024-11-06
Loosen LowestFee metric bound

jp1ac4:
The change in #29 led to the `LowestFee` metric lower bound being too tight, as found in the CI failure:
https://github.com/bitcoindevkit/coin-select/actions/runs/11584669281/job/32252247886

This PR tries to loosen the bound by assuming that no rounding up will be required when converting from weight units to vbytes, which would be the best-case scenario.

### DIFF
--- a/src/coin_selector.rs
+++ b/src/coin_selector.rs
@@ -200,12 +200,35 @@ impl<'a> CoinSelector<'a> {
             - self.implied_fee_from_feerate(target, drain.weights) as i64
     }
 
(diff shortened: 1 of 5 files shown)


jp1ac4:
Hi @LLFourn, would it be possible to release a new crate version with the changes from #29 and possibly this PR? Thank you.

LLFourn:
tACK af1333b3a13afccbfb96197c76d7dff72bf7818f

PR #35, opened 2025-09-15
Fix: allow zero-fee transactions in LowestFee metric

evanlinjin:
Change assertion from fee > 0 to fee >= 0 to handle valid zero-fee transactions. Add test case to verify the metric works correctly when target fee rate is zero.

Also make clippy happy and upgrade test dependencies.

### DIFF
--- a/tests/lowest_fee.rs
+++ b/tests/lowest_fee.rs
@@ -267,3 +267,55 @@ fn adding_another_input_to_remove_change() {
     assert!(score <= best_solution_score);
     assert_eq!(cs.selected_indices(), best_solution.selected_indices());
 }
+
+#[test]
+fn zero_fee_tx() {
+    let target_feerate = FeeRate::ZERO;
+    let long_term_feerate = FeeRate::DEFAULT_MIN_RELAY;
+
+    let target = Target {
+        fee: TargetFee {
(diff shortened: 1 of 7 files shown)


tests/lowest_fee.rs
```diff
@@ -267,3 +267,55 @@ fn adding_another_input_to_remove_change() {
     assert!(score <= best_solution_score);
     assert_eq!(cs.selected_indices(), best_solution.selected_indices());
 }
+
+#[test]
+fn zero_fee_tx() {
```
jp1ac4, line 272: Not sure if it matters, but AFAICT this test produces a solution with a non-zero fee (130 sats).

evanlinjin: If it's a solution with change then it's a problem. If it's without change, it's expected as we have a long-term feerate so that greatly disincentives the solution with change.

evanlinjin: But maybe we should check that a solution with change has 0 fee.

jp1ac4: Ah OK, the solution doesn't have change so that's fine. I wasn't sure if the intention of the test was to produce a transaction with zero fee.

PR #37, opened 2026-01-15
fix: add conditional FloatExt import for no_std builds

### LINKED ISSUE #36: Unable to build with no default features?
`cargo build --no-default-features` gave me a compile error. It seems that a `FloatExt` is not in scope.

```
error[E0599]: no method named `ceil` found for type `f32` in the current scope
  --> src/feerate.rs:86:35
   |
86 |         ((tx_weight as f32 / 4.0).ceil() * self.as_sat_vb()).ceil() as u64
   |                                   ^^^^ method not found in `f32`
   |
  ::: src/float.rs:67:8
   |
67 |     fn ceil(self) -> Self;
   |        ---- the method is available for `f32` here
   |
   = help: items from traits can only be used if the trait is in scope
help: trait `FloatExt` which provides `ceil` is implemented but not in scope; perhaps you want to import it
   |
 1 + use crate::float::FloatExt;
```

aagbotemi:
Fixes compilation error when building with `--no-default-features` by adding a conditional import of the `FloatExt` trait in `feerate.rs`.

* [x] I've signed all my commits
* [x] I ran `cargo fmt` and `cargo clippy` before committing
* [x] CHANGELOG.md updated

Fixes #36

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- Fix compilation error when building with `--no-default-features` (#36)
+
 # 0.4.0
 
 - Use `u64` for weights instead of u32
--- a/src/feerate.rs
+++ b/src/feerate.rs
@@ -1,3 +1,5 @@
+#[cfg(not(feature = "std"))]
+use crate::float::FloatExt;
 use crate::float::Ordf32;
 use core::ops::{Add, Sub};
 


evanlinjin (approved):
ACK 9becc092bfa5d4af078f689d74ffa6048a5666f0

Thanks for the fix!

PR #38, opened 2026-02-08
Integrate package context into fee and weight calculations; add absolute fee target

evanlinjin:
## Package context for CPFP

When a `Package` is set via `with_package()`, all fee and weight calculations now automatically include parent transaction context:

- `weight()` returns `child_weight + parent_weight`
- `fee()` returns `child_fee + parent_fee`
- `implied_feerate()` returns package feerate

New methods for when child-only values are needed:
- `weight_without_package()`
- `fee_without_package()`
- `package()` accessor

RBF calculations (`replacement_excess`) use child weight only since Bitcoin's RBF rule 4 applies to the replacing transaction, not the package.

## Absolute fee target

Adds a `TargetFee::absolute` field that acts as a minimum absolute fee floor, checked independently alongside the existing feerate and replacement constraints.

**Use cases:**
- **Payjoin (BIP 78):** The receiver must not decrease the original transaction's absolute fee when adding inputs.
- **Application-level policy minimums:** Supports `bdk_wallet`'s `TxBuilder::fee_absolute()` API where users set a fee floor regardless of feerate.

**Changes:**
- `TargetFee::absolute: u64` — minimum absolute fee the transaction must pay
- `CoinSelector::absolute_excess()` — how much the selection overshoots the absolute fee requirement
- `absolute_excess` integrated into `excess()` via `.min()` so the most binding constraint wins
- `implied_fee()` accounts for `absolute` via `.max()`
- `LowestFee::bound()` handles the absolute constraint for correct branch-and-bound behavior

Note: absolute fee is complementary to package context, not a replacement. Package context changes how the feerate constraint itself scales with child weight (the parent's fee/weight are inside the feerate calculation). An absolute fee floor is an independent constraint for cases where a fixed sat minimum comes from an external source.

### DIFF
--- a/src/coin_selector.rs
+++ b/src/coin_selector.rs
@@ -245,6 +245,7 @@ impl<'a> CoinSelector<'a> {
     /// this means the transaction will overpay for what it needs to reach `target`.
     pub fn excess(&self, target: Target, drain: Drain) -> i64 {
         self.rate_excess(target, drain)
+            .min(self.absolute_excess(target, drain))
             .min(self.replacement_excess(target, drain))
     }
 
@@ -259,7 +260,7 @@ impl<'a> CoinSelector<'a> {
     }
 
     /// How much the current selection overshoots the value need to satisfy `target.fee.rate` and
-    /// `target.value` (while ignoring `target.min_fee`).
+    /// `target.value` (while ignoring `target.fee.absolute`).
     pub fn rate_excess(&self, target: Target, drain: Drain) -> i64 {
         self.selected_value() as i64
             - target.value() as i64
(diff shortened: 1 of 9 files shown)


aagbotemi (commented):
The approach of adding `Package` to `CoinSelector` is good one and works well for the UTXO selection for CPFP use case.

evanlinjin:
I also want to add an absolute fee target.

**Payjoin (BIP 78):** The payjoin proposal MUST NOT decrease the absolute fee of the original transaction.

Some users may want to select their fee based on an absolute value (not sure why, but some wallets provide this feature).

src/coin_selector.rs
```diff
@@ -192,7 +260,7 @@ impl<'a> CoinSelector<'a> {
     }
 
     /// How much the current selection overshoots the value need to satisfy `target.fee.rate` and
-    /// `target.value` (while ignoring `target.min_fee`).
+    /// `target.value` (while ignoring `target.fee.absolute`).
     pub fn rate_excess(&self, target: Target, drain: Drain) -> i64 {
```
aagbotemi, line 264: `rate_excess` includes `parent_weight` in the required fee via `implied_fee_from_feerate()` -> `weight()` but doesn't include `parent_fee` in the actual fee, so the excess is `parent_fee` sats too low.

evanlinjin:
Closing in favor of #43.
Absolute fee target should just be a standalone PR.

PR #39, opened 2026-02-11
Fix tuple BnB metric ordering propagation

EliteCoder18:
Tuple BnB metrics currently propagates `requires_ordering_by_descending_value_pwu()` using `all()`. 

If one metric requires descending order and another does not, order is skipped, thus violating the requirement of the order-dependent metrics.

What this PR does:

1. Adds a test (`bnb_tuple_metric_respects_ordering_requirement`) demonstrating the issue.

2. The fix involves changing the ordering propagation to use any() instead of all().

The test is added in a separate commit to clearly demonstrate the failing behavior before the fix.

### DIFF
--- a/tests/bnb.rs
+++ b/tests/bnb.rs
@@ -52,6 +52,36 @@ impl BnbMetric for MinExcessThenWeight {
     }
 }
 
+#[derive(Clone, Copy)]
+struct OrderSensitive;
+
+impl BnbMetric for OrderSensitive {
+    fn score(&mut self, _cs: &CoinSelector<'_>) -> Option<Ordf32> {
+        Some(Ordf32(0.0))
+    }
+
+    fn bound(&mut self, _cs: &CoinSelector<'_>) -> Option<Ordf32> {
+        Some(Ordf32(0.0))
+    }
+
+    fn requires_ordering_by_descending_value_pwu(&self) -> bool {
+        true
+    }
+}
+
+#[derive(Clone, Copy)]
+struct NoOrderMetric;
+
+impl BnbMetric for NoOrderMetric {
+    fn score(&mut self, _cs: &CoinSelector<'_>) -> Option<Ordf32> {
+        Some(Ordf32(0.0))
+    }
+
+    fn bound(&mut self, _cs: &CoinSelector<'_>) -> Option<Ordf32> {
+        Some(Ordf32(0.0))
+    }
+}
+
 #[test]
 /// Detect regressions/improvements by making sure it always finds the solution in the same
 /// number of iterations.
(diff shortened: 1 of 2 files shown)


evanlinjin (changes requested):
Thanks for the fix! Before ACK/Merge, I propose the following changes:

### 1. Simplify tests

No need to actually run `.bnb_solutions`. We just need to make sure `((metric_a, 1.0), (metric_b, 1.0)).requires_ordering_by_descending_value_pwu()` returns the correct value.

* `score` and `bound` implementations of `OrderSensitive` and `NoOrderMetric` can just return `Some(Ord32(0.0))`.
* Have the following test cases:
  * Both metrics require ordering
  * One metric requires ordering
  * No metric requires ordering

### 2. Don't have a separate clippy/lint commit

Just merge the commits.

EliteCoder18:
Thanks for the suggestions. I've made the tests easier by only covering the three ordering situations and getting rid of the lint commit.

Please let me know if anything else needs to be changed

tests/bnb.rs
```diff
@@ -139,6 +169,21 @@ fn bnb_finds_solution_if_possible_in_n_iter() {
...
+        ((OrderSensitive, 1.0), (OrderSensitive, 1.0))
+            .requires_ordering_by_descending_value_pwu()
+    );
+    assert!(
+        ((OrderSensitive, 1.0), (NoOrderMetric, 1.0))
+            .requires_ordering_by_descending_value_pwu()
+    );
+    assert!(
+        !((NoOrderMetric, 1.0), (NoOrderMetric, 1.0)).requires_ordering_by_descending_value_pwu()
+    );
```
evanlinjin, line 187: **Nit:** For clarity.

```suggestion
    assert!(
        ((OrderSensitive, 1.0), (OrderSensitive, 1.0)).requires_ordering_by_descending_value_pwu(),
        "both requires ordering, so ordering needed"
    );
    assert!(
        ((OrderSensitive, 1.0), (NoOrderMetric, 1.0)).requires_ordering_by_descending_value_pwu(),
        "one requires ordering, so ordering needed"
    );
    assert!(
        !((NoOrderMetric, 1.0), (NoOrderMetric, 1.0)).requires_ordering_by_descending_value_pwu(),
        "none requires ordering, so no ordering needed"
    );
```

evanlinjin (commented):
Don't forget to do `cargo fmt` and `cargo clippy` otherwise CI will fail.

tests/bnb.rs
EliteCoder18: I appreciate your recommendation. I have included the assert messages to ensure clarity and have executed cargo fmt / cargo clippy locally. Implemented the revised modifications.

evanlinjin (approved):
ACK 66e82d475f0f38d02a7ac3dad6a5353767acbee4

PR #42, opened 2026-03-25
Add ancestor_bump_fee to Candidate

### LINKED ISSUE #24: CPFP: Allow telling coin selector that a candidate has unconfirmed parents

aagbotemi:
### Description

Add `ancestor_bump_fee` field to `Candidate`, allowing callers to
express the additional fee a child transaction must pay to bring
its unconfirmed ancestors up to the target feerate.

### Note to Reviewers
Cross-candidate ancestor deduplication is not handled at this layer.
Shared ancestors across candidates are counted per-candidate, which
may overcount when multiple selected candidates share an ancestor.
This is conservative (overpays, never underpays). The wallet layer
deduplicates post-selection.

Closes #24

### Changelog notice
- Added `ancestor_bump_fee` field to `Candidate` for ancestor-aware coin selection

#### All Submissions:

* [x] I followed the [contribution guidelines](https://github.com/bitcoindevkit/coin-select/blob/master/CONTRIBUTING.md)
* [x] I've added tests for the new feature
* [x] I've added docs for the new feature
* [x] I'm linking the issue being fixed by this PR

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,8 @@ let candidates = vec![
         weight: TR_KEYSPEND_TXIN_WEIGHT,
         // wether it's a segwit input. Needed so we know whether to include the
         // segwit header in total weight calculations.
-        is_segwit: true
+        is_segwit: true,
+        ancestor_bump_fee: 0
     },
     Candidate {
         // A candidate can represent multiple inputs in the case where you 
@@ -47,7 +48,8 @@ let candidates = vec![
         input_count: 2,
         weight: 2*TR_KEYSPEND_TXIN_WEIGHT,
         value: 3_000_000,
-        is_segwit: true
+        is_segwit: true,
+        ancestor_bump_fee: 0
     }
 ];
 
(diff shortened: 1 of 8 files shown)


evanlinjin (changes requested):
NACK

Thanks for working on this. However, this solution is flawed.

* `ancestor_bump_fee` is a function of the target feerate. This is incorrect as `Candidate` should only have fixed properties of an UTXO.
* This solution does not take ancestor overlaps into account. Please refer to https://github.com/bitcoindevkit/coin-select/issues/24#issuecomment-2650958507

Also, next time, please remember to reference pre-existing PRs addressing similar problems. This PR is essentially a competing solution to #38.

aagbotemi:
Thanks for the review. 


> - ancestor_bump_fee is a function of the target feerate. This is incorrect as Candidate should only have fixed properties of an UTXO.
> - This solution does not take ancestor overlaps into account. Please refer to https://github.com/bitcoindevkit/coin-select/issues/24#issuecomment-2650958507

Both points are valid.


> Also, next time, please remember to reference pre-existing PRs addressing similar problems. This PR is essentially a competing solution to https://github.com/bitcoindevkit/coin-select/pull/38.

I looked at #38 before opening this PR and concluded they addressed different layers. #38 operates at the transaction level with a single `Package` aggregate, this PR at the candidate level. I should have commented on #38 explaining this reasoning rather than opening a separate PR.

aagbotemi:
Closing in favor of #43

PR #45, opened 2026-05-17
test: proptest that LowestFee finds planted single-input optimum

evanlinjin:
## Summary

Adds a constructive proptest covering an edge case that `can_eventually_find_best_solution` doesn't reach: BnB at large N (up to ~200 candidates) finding an obvious 1-input optimum without paying the cost of exhaustive search.

The construction:

- Plant one candidate at index 0 with `value = target_value + fee_for_1_input` (computed by probing with a dummy candidate's weight).
- Surround it with up to 200 dust candidates (value=1, P2WPKH-like weight).
- Since (a) `n_noise ≤ 200 < target_value ≥ 10_000`, noise can never sum to cover the target, and (b) adding any noise to the planted candidate either makes the fee delta exceed the +1 sat value (target no longer met) or triggers drain (score strictly > `fee_for_1_input`), the planted candidate alone is the unique optimum.
- BnB is expected to converge within `10 * (n_noise + 2)` rounds.

No production code changes — this is a test-only addition.

## Test plan

- [x] `cargo test --release --test lowest_fee lowest_fee_finds_planted_single_input` (256 proptest cases)
- [x] `cargo test --release --test lowest_fee` (full suite, all 8 tests pass)

### DIFF
--- a/tests/lowest_fee.rs
+++ b/tests/lowest_fee.rs
@@ -4,7 +4,7 @@ mod common;
 use bdk_coin_select::metrics::{Changeless, LowestFee};
 use bdk_coin_select::{
     BnbMetric, Candidate, ChangePolicy, CoinSelector, Drain, DrainWeights, FeeRate, Replace,
-    Target, TargetFee, TargetOutputs, TX_FIXED_FIELD_WEIGHT,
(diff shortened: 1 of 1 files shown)


evanlinjin:
No longer needed.

PR #48, opened 2026-06-25
Add MaxWeight to cap transaction weight during selection

aagbotemi:
### Description

`max_weight` field added to `Target`, making the maximum transaction weight a selection constraint enforced in `is_target_met`.

### Notes for review
- The weight cap lives on Target and flows through `is_target_met`/`is_target_met_with_drain`.
- `select_until_target_met` now returns `SelectError { InsufficientFunds, MaxWeightExceeded }` so a weight cap failure isn't mislabeled as a funds shortfall.

### Changelog notice
- Added `max_weight` field to `Target`
- `select_until_target_met` returns `SelectError`

### DIFF
--- a/src/coin_selector.rs
+++ b/src/coin_selector.rs
@@ -308,7 +308,7 @@ impl<'a> CoinSelector<'a> {
         self.selected_value() as i64 - (self.input_weight() as f32 * feerate.spwu()).ceil() as i64
     }
 
-    // /// Waste sum of all selected inputs.
+    /// Waste sum of all selected inputs.
     fn input_waste(&self, feerate: FeeRate, long_term_feerate: FeeRate) -> f32 {
         self.input_weight() as f32 * (feerate.spwu() - long_term_feerate.spwu())
     }
@@ -429,15 +429,20 @@ impl<'a> CoinSelector<'a> {
         self.unselected_indices().next().is_none()
     }
 
+    /// Whether the tx implied by the current selection and `drain` is within `target.max_weight`.
+    fn is_within_max_weight(&self, target: Target, drain: Drain) -> bool {
+        match target.max_weight {
+            Some(max_weight) => self.weight(target.outputs, drain.weights) <= max_weight,
+            None => true,
+        }
+    }
+
     /// Whether the constraints of `Target` have been met if we include a specific `drain` ouput.
     ///
-    /// Note if [`is_target_met`] is true and the `drain` is produced from the [`drain`] method then
-    /// this method will also always be true.
-    ///
-    /// [`is_target_met`]: Self::is_target_met
-    /// [`drain`]: Self::drain
+    /// Note: even if [`Self::is_target_met`] is true, this can be false when adding the `drain` pushes
+    /// the transaction weight over [`Target::max_weight`].
     pub fn is_target_met_with_drain(&self, target: Target, drain: Drain) -> bool {
-        self.excess(target, drain) >= 0
+        self.excess(target, drain) >= 0 && self.is_within_max_weight(target, drain)
     }
 
     /// Whether the constraints of `Target` have been met.
@@ -467,16 +472,14 @@ impl<'a> CoinSelector<'a> {
             },
         );
         if excess > change_policy.min_value as i64 {
-            debug_assert_eq!(
-                self.is_target_met(target),
-                self.is_target_met_with_drain(
-                    target,
-                    Drain {
-                        weights: change_policy.drain_weights,
-                        value: excess as u64
-                    }
-                ),
-                "if the target is met without a drain it must be met after adding the drain"
+            let drain = Drain {
+                weights: change_policy.drain_weights,
+                value: excess as u64,
+            };
+            debug_assert!(
+                self.is_target_met_with_drain(target, drain)
+                    || !self.is_within_max_weight(target, drain),
+                    "if the target is met without a drain it must be met after adding the drain, unless max weight is exceeded"
             );
             Some(excess as u64)
         } else {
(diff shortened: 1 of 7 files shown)


murchandamus (commented):
I’m not actively contributing to this repository, so please feel free to ignore me. I noticed that this PR only adds `MaxWeight` for BnB, but a maximum weight would be especially relevant in the context of creating [TRUC transactions](https://github.com/bitcoin/bips/blob/master/bip-0431.mediawiki) that are limited to 10,000 vB or 1000 vB respectively, so it would be relevant to all appropriate for all coinselection algorithms.

evanlinjin:
@murchandamus Thank you for the insight and you have a point. Maybe the best move is to make `MaxWeight` part of the `Target` as it is a selection constraint.

This will also make bitcoindevkit/bdk-tx#47 implementation cleaner.

aagbotemi:
> I’m not actively contributing to this repository, so please feel free to ignore me. I noticed that this PR only adds MaxWeight for BnB, but a maximum weight would be especially relevant in the context of creating [TRUC transactions](https://github.com/bitcoin/bips/blob/master/bip-0431.mediawiki) that are limited to 10,000 vB or 1000 vB respectively, so it would be relevant to all appropriate for all coinselection algorithms.

Thanks @murchandamus for the review.

> Maybe the best move is to make MaxWeight part of the Target as it is a selection constraint.
> 
> This will also make https://github.com/bitcoindevkit/bdk-tx/pull/47 implementation cleaner.


Thanks @evanlinjin, I took your suggestion and moved the cap onto Target as a selection constraint.

src/coin_selector.rs
```diff
@@ -467,17 +472,6 @@ impl<'a> CoinSelector<'a> {
...
-                self.is_target_met(target),
-                self.is_target_met_with_drain(
-                    target,
-                    Drain {
-                        weights: change_policy.drain_weights,
-                        value: excess as u64
-                    }
-                ),
-                "if the target is met without a drain it must be met after adding the drain"
-            );
```
evanlinjin, line 480: What's the motivation to remove this?

src/coin_selector.rs
```diff
@@ -520,13 +514,26 @@ impl<'a> CoinSelector<'a> {
...
+    /// - [`SelectError::MaxWeightExceeded`] if the value is met but the selection exceeds
+    ///   [`Target::max_weight`].
+    pub fn select_until_target_met(&mut self, target: Target) -> Result<(), SelectError> {
         self.select_until(|cs| cs.is_target_met(target))
-            .ok_or_else(|| InsufficientFunds {
-                missing: self.excess(target, Drain::NONE).unsigned_abs(),
+            .ok_or_else(|| {
+                let excess = self.excess(target, Drain::NONE);
+                // excess < 0 means unfundable (InsufficientFunds),
+                // else value is met so the weight cap failed (MaxWeightExceeded).
```
evanlinjin: What happens if there is a changeless solution?

src/coin_selector.rs
aagbotemi: I intended to use `excess`. Now that `max_weight` is on `Target`, `is_target_met` and `is_target_met_with_drain` both check the cap, and they can differ on weight. The replacement check only the value it was meant to guard, independent of weight.

src/coin_selector.rs
aagbotemi: `excess` is measured with `Drain::NONE`, so a changeless solution has `excess >= 0`, meaning value is met, so the failure is reported as `MaxWeightExceeded`, not `InsufficientFunds`.

evanlinjin: Sorry, I misunderstood the code - this logic is for deciding the error variant after we already know that target is not met.

src/coin_selector.rs
```diff
@@ -520,13 +524,27 @@ impl<'a> CoinSelector<'a> {
...
-    pub fn select_until_target_met(&mut self, target: Target) -> Result<(), InsufficientFunds> {
+    /// - [`SelectError::InsufficientFunds`] if the candidates can't cover the target value.
+    /// - [`SelectError::MaxWeightExceeded`] if the value is met but the selection exceeds
+    ///   [`Target::max_weight`].
+    pub fn select_until_target_met(&mut self, target: Target) -> Result<(), SelectError> {
         self.select_until(|cs| cs.is_target_met(target))
-            .ok_or_else(|| InsufficientFunds {
-                missing: self.excess(target, Drain::NONE).unsigned_abs(),
+            .ok_or_else(|| {
+                let excess = self.excess(target, Drain::NONE);
```
evanlinjin: Excess can be negative because inputs are low in effective value and high in weight -- in which case, we should return `MaxWeightExceeded`.

src/coin_selector.rs
```diff
@@ -467,16 +472,15 @@ impl<'a> CoinSelector<'a> {
...
                     Drain {
                         weights: change_policy.drain_weights,
-                        value: excess as u64
-                    }
-                ),
-                "if the target is met without a drain it must be met after adding the drain"
+                        value: excess as u64,
+                    },
+                ) >= 0,
+                "if the target value is met without a drain it must be met after adding the drain"
```
evanlinjin: Shouldn't the check be: if the target is met without a drain, it must be met after adding the drain, unless if max weight is exceeded?

src/coin_selector.rs
evanlinjin: In this case, it's better to change the check rather than remove it.

aagbotemi: Yes, I have changed the check

src/coin_selector.rs
aagbotemi: Thanks for the catch. I've fixed it.

src/coin_selector.rs
evanlinjin: This is wrong. Adding a drain output increases the weight of the transaction so we cannot guarantee that it'll be within our max weight target (as the `debug_assert` implies).

aagbotemi: Good. We can either keep a value-only check (`excess(target, drain) >= 0`, scoped to value only) or drop the assert, since with `drain.value = excess` here it's always true. Open to your thoughts.

evanlinjin:
We need a bit more thought looking at how adding `max_weight` to `Target` changes that other parts of the codebase.

* The logic of `is_selection_possible` is now wrong. Previously, `Target` was a purely monotone predicate -- it's fields calculated excess which can monotonously be increased by adding more effective inputs. Putting `max_weight` in `Target` changed this property as `max_weight` puts an upper bound on the weights of the inputs added. To "fix" it, we need to do a more expensive knapsack algorithm -- is this the best move?

* How does the `max_weight` parameter affect the `LowestFee` metric? I've noticed that proptests are not updated to test with different `max_weight` values.

evanlinjin:
@aagbotemi I'm going to have a go at finding the right solution. The problem is a lot more complex than initially thought.

aagbotemi:
I went down some rabbit hole. Making `is_selection_possible` exact under a weight cap turns it into a knapsack-style problem where we have to search subsets. Any exact approach is too heavy for a simple pre-check, and BnB already does that kind of search. The idea I was working with earlier is to have `is_selection_possible` ignore `max_weight`. Then `false` still means truly impossible, `true` means possible ignoring the cap, and selection itself reports `MaxWeightExceeded`. I also wrote a proptest that brute-forces every subset to verify any solution we pick [here](https://github.com/bitcoindevkit/coin-select/compare/master...aagbotemi:coin-select:feat/max-weight-bound-fix). Interested to hear findings.

evanlinjin:
@aagbotemi I opened #51 as a replacement. The core issue is that I believe this PR's implementation is incorrect and the tests don't catch it because `LowestFee` BnB is never actually run with `max_weight` set.

It turns out that the value and weight predicates should be considered separately to avoid BnB pruning too aggressively and basing #51 on #49 really simplified the implementation.

evanlinjin:
Closing as #51 is merged.

PR #50, opened 2026-07-02
feat: add CoinSelector::select_srd (Single Random Draw)

evanlinjin:
## Motivation

Adds **Single Random Draw** coin selection — a port of Bitcoin Core's `SelectCoinsSRD`. Unlike `run_bnb` with `LowestFee` (which minimizes fees), SRD deliberately aims for a **healthy-sized, privacy-friendly change output**, avoiding tiny "toxic" change. It's the natural complement to the fee-minimizing metric from #49.

## What it does

`CoinSelector::select_srd` adds candidates in random order until the change reaches `change_lower`, then returns the `Drain` to attach (or a `SelectError`).

```rust
pub fn select_srd(
    &mut self,
    target: Target,
    drain_weights: DrainWeights,
    change_lower: u64,
    rng: impl FnMut() -> u64,
) -> Result<Drain, SelectError>
```

- **Fixed lower bound, like Core.** We stop as soon as the change reaches `change_lower`; the change *amount* comes out random on its own, because candidates are added in random order and the final change is wherever the last random input pushed it. Core uses the same fixed-threshold approach (and its comment notes SRD "will result in a random change amount anyway"), so there's no need to randomize the target.
- The stop condition reuses `excess(target, drain_weights)`, so there's **no new fee math** — it inherits the existing rate/absolute/replacement handling.
- **Zero new dependencies**: RNG is a plain `impl FnMut() -> u64` (the crate is deliberately dependency-free), used to shuffle the candidates.
- Also adds **`CoinSelector::shuffle_candidates`**, a public Fisher-Yates shuffle of the candidate order driven by the same `impl FnMut() -> u64` RNG. `select_srd` is built on it, and it's useful on its own for randomizing selection order.
- `CHANGE_LOWER` (Core's 50k) is added as the sensible default for `change_lower`.
- Already-selected candidates are kept and counted toward the target.

`select_srd` returns `SelectError::InsufficientFunds` when the candidates can't cover target + `change_lower`, and `SelectError::MaxWeightExceeded` when a `Target::max_weight` cap is set and reaching `change_lower` would trip it. Recovering from the weight cap by evicting the least-valuable inputs (Core's `max_selection_weight`) is left as a documented `// TODO` pending the max-weight PR — for now it errors rather than evicts.

## Tests

`tests/srd.rs` covers: change ≥ `change_lower` and target met across 300 seeds; a custom `change_lower` is honored; insufficient funds (asserting the `InsufficientFunds` variant); `MaxWeightExceeded` when a weight cap is tripped; per-seed determinism; preselected inputs preserved. Full suite + `no_std` build + clippy are green.


src/coin_selector.rs
```diff
@@ -548,6 +552,70 @@ impl<'a> CoinSelector<'a> {
...
+
+        // Shuffle the unselected candidates (Fisher-Yates) and add them until we have enough change.
+        let mut candidates: Vec<usize> = self.unselected_indices().collect();
+        for i in (1..candidates.len()).rev() {
+            let j = (rng() % (i as u64 + 1)) as usize;
+            candidates.swap(i, j);
+        }
+        for index in candidates {
+            if self.excess(target, drain) >= change_lower as i64 {
+                break;
```
evanlinjin: Why don't we just return here instead of checking excess again?

For the error case, call the `self.select` first in the loop and store the `self.excess` as a `let mut` outside the loop to use for `InsufficientFunds`.

evanlinjin: Done in the latest push. Now it selects first, returns `Ok` from inside the loop, and keeps `change` (the excess) in a `let mut` so the `InsufficientFunds` case reuses it instead of recomputing. I kept a guard before the loop for the case where already-selected inputs are enough, so it doesn't select an unnecessary extra input.

evanlinjin: Follow-up: after a bit more back-and-forth this landed as a `loop` + `break value` (the "loop-and-a-half"), which is the cleanest form here — one `Ok`, one `Err`, no post-loop re-check, no over-selection, and `change` (the excess) is never recomputed for the result:

```rust
let mut candidates = candidates.into_iter();
let mut change = self.excess(target, drain);
loop {
    if change >= change_lower as i64 {
        break Ok(Drain { weights: drain_weights, value: change as u64 });
    }
    match candidates.next() {
        Some(index) => {
            self.select(index);
            change = self.excess(target, drain);
        }
        None => break Err(InsufficientFunds {
            missing: (change_lower as i64 - change).unsigned_abs(),
        }),
    }
}
```

The top check catches all three success paths (already-sufficient, met mid-loop, met on the final candidate); `None` is the single failure exit. In `1381fd4`.

evanlinjin (commented):
ACK 7ab63772a018dbd6c1e51f672fc299aa16db5602

src/coin_selector.rs
```diff
@@ -4,6 +4,10 @@ use crate::float::FloatExt;
 use crate::{bitset::Bitset, bnb::BnbMetric, float::Ordf32, ChangePolicy, FeeRate, Target};
 use alloc::{sync::Arc, vec::Vec};
 
+/// The minimum change amount Bitcoin Core's `SelectCoinsSRD` targets; a sensible default for the
+/// `change_lower` argument of [`CoinSelector::select_srd`].
+pub const CHANGE_LOWER: u64 = 50_000;
```
yancyribbens, line 9: It looks like this const `CHANGE_LOWER` is only used by the tests?  If so, should it just be moved to the tests module?

evanlinjin: This is also useful as an input for `select_srd`.

yancyribbens: ah, so the user of the interface would pass this as `change_lower` param?  Anyway, I wonder if change_lower is even that useful since the caller could instead just increase the target amount by that much instead of needing this param at all.

evanlinjin: Target is for defining the feerate, minimum absolute fee and recipient outputs. `change_lower` is defining the minimum amount for the change output.

PR #55, opened 2026-07-07
feat!: split NoBnbSolution into infeasible vs round-limit variants

evanlinjin:
## Description

`run_bnb`'s failure is currently a single `NoBnbSolution { max_rounds, rounds }`, which conflates three different situations. This makes it an enum:

- `InsufficientFunds` — the candidates can't cover the target value.
- `MaxWeightExceeded` — the value is coverable, but every selection busts `max_weight` (mirrors the `SelectError` from `select_until_target_met`).
- `RoundLimit { max_rounds, rounds }` — hit the round cap; a solution may still exist with more rounds.

## How — without touching `BnbMetric`

The metric returning `None` just means "prune"; it never needs an error channel. All the information already lives in `run_bnb`: if the search runs to completion (`rounds < max_rounds`, or a final pull drains the iterator) then the tree was fully explored and no selection satisfies the target — a genuine infeasibility. A cheap `is_fundable` value check then splits that into value (`InsufficientFunds`) vs weight (`MaxWeightExceeded`). If instead we stopped on the round cap, it's `RoundLimit`.

The `MaxWeightExceeded` attribution is sound for the crate's metrics: a cap-blind metric can only exhaust-empty when the value itself is unreachable (otherwise it would have *returned* an over-cap selection, not failed), so you never falsely land there.

## Breaking

`NoBnbSolution` becomes an enum; the `max_rounds`/`rounds` fields move under the `RoundLimit` variant. `Display`/`Error` still hold.

## Tests

Deterministic tests for each variant: `run_bnb_reports_{insufficient_funds,max_weight_exceeded,round_limit}`.

### DIFF
--- a/src/coin_selector.rs
+++ b/src/coin_selector.rs
@@ -636,10 +636,24 @@ impl<'a> CoinSelector<'a> {
             .inspect(|_| rounds += 1)
             .flatten()
             .last();
-        let (selector, score) = best.ok_or(NoBnbSolution { max_rounds, rounds })?;
-        let drain = iter.metric.drain(&selector, target);
-        *self = selector;
-        Ok((score, drain))
+        if let Some((selector, score)) = best {
+            let drain = iter.metric.drain(&selector, target);
+            *self = selector;
+            return Ok((score, drain));
+        }
+
+        // No solution. If the iterator still has an item we stopped at the round limit and a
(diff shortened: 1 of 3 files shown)


LLFourn:
conceptACK

PR #59, opened 2026-08-04
Give `CoinSelector` its target instead of threading it through every call

evanlinjin:
## Why

A `CoinSelector` is built for one target and evaluated against it throughout — yet every method took the target as a parameter. Nothing stopped `cs.excess(target_a, drain)` being followed by `cs.is_funded(target_b)`, and the metrics' correctness arguments (e.g. `LowestFee::bound`'s proof that a changeless superset always costs more) are all stated at a *fixed* target, held together by convention rather than by types.

The crate had already reached this conclusion one layer down: `BnbIter` stored the target as a field, took it once in `new`, and re-passed it into `metric.score`/`metric.bound` at every node. This moves the binding up to where it belongs and deletes the re-threading.

It also unblocks follow-up work: with the selector knowing its target feerate, ancestor-aware CPFP pricing (#24) can be derived internally at the right rate instead of being validated at every call site. That branch is based on this one.

## What

`CoinSelector::new(candidates, target)` owns the target; `CoinSelector::target()` exposes it for metrics that read it. Twenty signatures **lose** a parameter: fifteen public methods (`excess`, `implied_fee`, `is_funded`, `drain`, `select_until_target_met`, the four `*_excess`, …), `bnb_solutions`/`run_bnb`, and all three `BnbMetric` methods. Across the existing tests and benches, no selector was ever evaluated against more than one target — the per-call flexibility had no consumer.

## Breaking changes

External `BnbMetric` implementations drop the `target: Target` parameter:

```rust
fn score(&mut self, cs: &CoinSelector<'_>) -> Option<Ordf32>;
fn bound(&mut self, cs: &CoinSelector<'_>) -> Option<Ordf32>;
fn drain(&mut self, cs: &CoinSelector<'_>) -> Drain;
```

Call sites move the target from each method call to `CoinSelector::new`. To evaluate a second target, build a second selector.

### DIFF
--- a/src/bnb.rs
+++ b/src/bnb.rs
@@ -1,6 +1,6 @@
 use core::cmp::Reverse;
 
-use crate::{float::Ordf32, Drain, Target};
+use crate::{float::Ordf32, Drain};
 
 use super::CoinSelector;
 use alloc::collections::BinaryHeap;
@@ -11,8 +11,6 @@ use alloc::collections::BinaryHeap;
 pub(crate) struct BnbIter<'a, M: BnbMetric> {
     queue: BinaryHeap<Branch<'a>>,
     best: Option<Ordf32>,
-    /// The target the metric scores selections against.
-    pub(crate) target: Target,
     /// The `BnBMetric` that will score each selection
     pub(crate) metric: M,
 }
@@ -55,7 +53,7 @@ impl<'a, M: BnbMetric> Iterator for BnbIter<'a, M> {
 
         let mut return_val = None;
         if !branch.is_exclusion {
-            if let Some(score) = self.metric.score(&selector, self.target) {
+            if let Some(score) = self.metric.score(&selector) {
                 let better = match self.best {
                     Some(best_score) => score < best_score,
                     None => true,
@@ -73,11 +71,10 @@ impl<'a, M: BnbMetric> Iterator for BnbIter<'a, M> {
 }
 
 impl<'a, M: BnbMetric> BnbIter<'a, M> {
-    pub(crate) fn new(mut selector: CoinSelector<'a>, target: Target, metric: M) -> Self {
+    pub(crate) fn new(mut selector: CoinSelector<'a>, metric: M) -> Self {
         let mut iter = BnbIter {
             queue: BinaryHeap::default(),
             best: None,
-            target,
             metric,
         };
 
@@ -91,7 +88,7 @@ impl<'a, M: BnbMetric> BnbIter<'a, M> {
     }
 
     fn consider_adding_to_queue(&mut self, cs: &CoinSelector<'a>, is_exclusion: bool) {
-        let bound = self.metric.bound(cs, self.target);
+        let bound = self.metric.bound(cs);
         if let Some(bound) = bound {
             let is_good_enough = match self.best {
                 Some(best) => best > bound,
@@ -205,7 +202,7 @@ pub trait BnbMetric {
     /// Get the score of a given selection for `target`.
     ///
     /// If this returns `None`, the selection is invalid.
-    fn score(&mut self, cs: &CoinSelector<'_>, target: Target) -> Option<Ordf32>;
+    fn score(&mut self, cs: &CoinSelector<'_>) -> Option<Ordf32>;
 
     /// Get the lower bound score using a heuristic for `target`.
     ///
@@ -214,13 +211,13 @@ pub trait BnbMetric {
     ///
     /// If this returns `None`, the current branch and all descendant branches will not have valid
     /// solutions.
-    fn bound(&mut self, cs: &CoinSelector<'_>, target: Target) -> Option<Ordf32>;
+    fn bound(&mut self, cs: &CoinSelector<'_>) -> Option<Ordf32>;
 
     /// The change output (a.k.a. drain) this metric decides on for the given selection and `target`,
     /// or [`Drain::NONE`] if it decides there should be no change.
     ///
     /// Call this on a branch-and-bound solution to get the change output the metric optimized against.
-    fn drain(&mut self, cs: &CoinSelector<'_>, target: Target) -> Drain;
+    fn drain(&mut self, cs: &CoinSelector<'_>) -> Drain;
 
     /// Returns whether the metric requies we order candidates by descending value per weight unit.
     fn requires_ordering_by_descending_value_pwu(&self) -> bool {
--- a/src/coin_selector.rs
+++ b/src/coin_selector.rs
@@ -18,6 +18,7 @@ pub const CHANGE_LOWER: u64 = 50_000;
 #[derive(Debug, Clone)]
 pub struct CoinSelector<'a> {
     candidates: &'a [Candidate],
+    target: Target,
     selected: Bitset,
     banned: Bitset,
     candidate_order: Arc<Vec<usize>>,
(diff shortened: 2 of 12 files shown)


src/metrics/lowest_fee.rs
```diff
@@ -71,17 +68,15 @@ impl LowestFee {
...
-            .map_or(Drain::NONE, |value| Drain {
-                weights: self.drain_weights,
-                value,
-            });
-        let fee_for_the_tx = cs.fee(target.value(), drain.value);
+        let drain = self.drain_value(cs).map_or(Drain::NONE, |value| Drain {
+            weights: self.drain_weights,
+            value,
+        });
+        let fee_for_the_tx = cs.fee(cs.target().value(), drain.value);
```
LLFourn: Non-blocking, API consistency: `fee(target_value, drain_value)`, `weight(target_outputs, drain_weights)` and `implied_feerate(target_outputs, drain)` still take a piece of the target as an argument. They never read the stored target, so they can't mix targets the way `excess`/`is_funded` could before, and they don't affect either bound. But every caller in the crate, tests and README passes the selector's own target: this line, `lowest_fee.rs` L165/L257 (`cs.weight(cs.target().outputs, …)`), the five `self.weight(self.target.outputs, …)` calls in `coin_selector.rs`, `tests/common.rs` L81/L121 and `tests/srd.rs` L133. `tests/weight.rs` now builds a `Target` from `target_ouputs` and then passes `target_ouputs` to `weight`/`implied_feerate` again.

Since this PR already breaks these call sites, consider making the common case read like the rest of the API, e.g. `cs.fee(drain.value)` / `cs.weight(drain_weights)` / `cs.implied_feerate(drain)`. If you want to keep the "what if the outputs were different" calculation, keep it under a separate name. Happy to leave this for a follow-up if you prefer.

src/bnb.rs
```diff
@@ -205,7 +202,7 @@ pub trait BnbMetric {
     /// Get the score of a given selection for `target`.
```
LLFourn: These three docs still describe `target` as if it were a parameter ("for `target`", "the given selection and `target`"). Suggest "Get the score of the selection `cs` against `cs.target()`", and the same for `bound` and `drain`.

It would also help to say here, where implementers will look, why the target is on the selector. `bound` has to be a lower bound on every descendant's `score`, and that argument only holds because `BnbIter` clones the one selector. So every node is guaranteed to carry the same target.

Other stale mentions in `coin_selector.rs`: `implied_fee`'s doc still says "to satisfy `target_fee`" (L283), `drain_value` says "constraints of `target`" (L514), and `excess`/`is_fundable`/`select_until_target_met` refer to "`target`" as if the caller supplied it (L128, L195, L198, L574).

src/coin_selector.rs
```diff
@@ -35,15 +36,24 @@ impl<'a> CoinSelector<'a> {
...
             candidates,
+            target,
             selected: Bitset::with_capacity(candidates.len()),
             banned: Bitset::with_capacity(candidates.len()),
             candidate_order: Arc::new((0..candidates.len()).collect::<Vec<_>>()),
         }
     }
 
+    /// What this selector is funding.
+    pub fn target(&self) -> Target {
```
LLFourn, line 53: The PR description says the replacement for evaluating a second target is "build a second selector". But that loses the current selection, the bans and the candidate order (`candidate_order` is private). To check an existing selection against a higher feerate, for example when deciding whether a fee bump needs more inputs, a caller now has to replay `selected_indices()`, `ban` and `sort_*` by hand.

A way to derive a new selector keeps the one-target-per-selector invariant while restoring that ability:

```rust
/// A copy of this selector (selection, bans and candidate order) that funds `target` instead.
pub fn with_target(&self, target: Target) -> CoinSelector<'a> {
    CoinSelector { target, ..self.clone() }
}
```

If you'd rather not add it, please at least put the "build a new selector" guidance in the CHANGELOG entry, so migrating callers know the selection doesn't carry over.

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Unreleased
 
-- **Breaking:** `BnbMetric`'s `score`, `bound`, and `drain` take the `target: Target` as a parameter, and `CoinSelector::run_bnb`/`bnb_solutions` gain a leading `target` argument. Consequently `LowestFee` and `Changeless` no longer store a `target` field. This removes the target that `Changeless<M>` previously had to keep in sync with its inner metric, and aligns the metric API with the rest of `CoinSelector`, where `target` is always passed in.
+- **Breaking:** `CoinSelector` now owns its `Target`. `CoinSelector::new(candidates, target)` takes it and `CoinSelector::target()` returns it, and it is fixed for the life of the selector. Every method that took a `target: Target` argument no longer does, including `excess`, `missing`, `rate_excess`, `implied_fee`, `is_funded`, `is_within_max_weight`, `drain`, `select_until_target_met`, `select_srd`, `run_bnb`, and `bnb_solutions`. `BnbMetric`'s `score`, `bound`, and `drain` read the target from the `CoinSelector` they are given, so `LowestFee` and `Changeless` no longer store a `target` field. This removes the target that `Changeless<M>` previously had to keep in sync with its inner metric.
```
LLFourn: Two small additions would help people migrating:
- Give the new `BnbMetric` signatures (`fn score(&mut self, cs: &CoinSelector<'_>) -> Option<Ordf32>` etc.) here, as the PR description does. External implementers will look in the CHANGELOG, not the PR.
- Say that the way to evaluate another target is a new selector, and that the current selection does not carry over (or point to `with_target` if you add it).

LLFourn (approved):

src/metrics/lowest_fee.rs
evanlinjin: Done — since the PR breaks these call sites anyway, they now read like the rest of the API:

- `weight(drain_weights)` and `implied_feerate(drain)` no longer take `TargetOutputs`
- `fee(drain_value)` no longer takes `target_value`
- and while in there, `effective_value()` and `select_all_effective()` no longer take a `FeeRate` — every caller passed the selector's own target rate too (the private `input_waste` likewise)

No separate "what if the outputs were different" variant: nothing in the crate or the tests wanted one. `SelectionProblem::with_target` + `CoinSelector::with_problem` (added downstream in #85, see the other thread) covers the real case of measuring against a second target.

src/bnb.rs
evanlinjin: Fixed. `score`/`bound`/`drain` now say "the selection `cs` against `cs.target()`", and the trait doc carries the invariant:

> Every selection passed to these methods carries its own `Target`, reachable via `CoinSelector::target`. `bound` is only a valid lower bound on the `score` of the descendants of `cs` because the search clones `cs` to build them, so every node of a single search is guaranteed to have the same target.

The other stale mentions are reworded too: `implied_fee` ("the `target`'s `TargetFee`"), `drain_value`, `excess`, `is_fundable` and `select_until_target_met` now all point at `Self::target` rather than an argument that no longer exists.

src/coin_selector.rs
evanlinjin: Good catch — replaying `selected_indices()`, `ban` and the sort by hand is not a migration path.

Added, though the shape differs slightly by the end of the stack. In this PR the target sits on the selector, so `with_target` is exactly what you wrote. Downstream (#85) the target moves into a `SelectionProblem` that the selector borrows, so it splits in two:

```rust
let bumped_problem = problem.with_target(higher_feerate_target);
let bumped = selector.with_problem(&bumped_problem);
assert_eq!(bumped.selected_indices(), selector.selected_indices());
```

`with_target` clones the problem with a new target (keeping candidates and, in #85, ancestry); `with_problem` carries the selection, the bans and the candidate order across, and asserts the candidate counts match since the selection is by index. There's a doctest on `with_problem` using your fee-bump example. The CHANGELOG entry mentions it too.

CHANGELOG.md
evanlinjin: Both added to the entry: the new `BnbMetric` signatures spelled out in full, and the note that another target means `SelectionProblem::with_target` + `CoinSelector::with_problem` (the selection *does* carry over through those), while `CoinSelector::new` starts empty.

I also went over the CHANGELOG across the whole stack while I was in there, since the entry gets amended by #63, #84 and #85 in turn.

noahjoeris (approved):
ACK 57114a8e04b5b0b3ae65c720635f2be5b08bd0d6

PR #53, opened 2026-07-02
perf!: delta-aware BnbMetric via SelectionView/compute_view

evanlinjin:
## Summary

Makes the branch-and-bound metric evaluator **delta-aware**, eliminating the dominant cost in `run_bnb`. A flamegraph showed ~65% of `run_bnb_lowest_fee` time was spent in `cs.selected_value()` (47.5%) and `cs.input_weight()` (17.3%) — both `O(|selected|)` walks recomputed many times per branch via `excess`/`is_target_met`/`drain`.

Now BnB maintains a per-branch `SelectionCache` (running aggregates: `value_sum`, `weight_sum`, `input_count`, `segwit_count`, `candidate_count`). Inclusion expansions call `cache.add(c)` — `O(1)`. Metrics receive a `&SelectionView<'_>` (a handle over `&CoinSelector` + `Cow<SelectionCache>`) whose read methods (`selected_value`, `input_weight`, `excess`, `is_target_met`, `drain`, …) are all `O(1)`.

Outside the hot path, `cs.compute_view()` builds a fresh cache once on demand, collapsing the ~15 duplicated `O(|selected|)` read methods that previously lived on `CoinSelector` into a single entry point.

## Benchmarks

Criterion median; delta-aware vs the bitset baseline, vs the pre-bitset master:

| bench                    | master  | bitset  | delta-aware | total speedup |
|--------------------------|---------|---------|-------------|---------------|
| clone/4096               | 5.68 µs | 58 ns   | 52 ns       | 109×          |
| run_bnb_lowest_fee/20    | 470 µs  | 417 µs  | 158 µs      | 3.0×          |
| run_bnb_lowest_fee/50    | 26.0 ms | 10.3 ms | 5.4 ms      | 4.8×          |
| run_bnb_lowest_fee/100   | 73.2 ms | 40.2 ms | 14.2 ms     | 5.2×          |
| run_bnb_lowest_fee/200   | 1.27 s  | 1.10 s  | 196 ms      | 6.5×          |

The gain is largest at large `n`, where `selected_value`/`input_weight` recomputation dominated. (The n=200 row measures the round-capped exploration path — see the bench commit for the explicit solution-finding vs cap-exhaustion group split.)

## Commits

Each commit independently passes the full CI gate (fmt, clippy, doc `-D warnings`, `--release` tests, `--no-default-features` build), so the branch is bisectable.

[...]

PR #64, opened 2026-08-12
feat!: ancestor-aware coin selection (CPFP bump via SelectionProblem)

### LINKED ISSUE #24: CPFP: Allow telling coin selector that a candidate has unconfirmed parents

### LINKED ISSUE #65: LowestFee: tighten the BnB bound when candidates have unconfirmed ancestors
## Background

Branch and bound (BnB) walks a tree of possible coin selections. At each node it asks the metric for a **bound**: the lowest score anything in this part of the tree could possibly reach. If that bound is already worse than the best answer found so far, the whole branch is skipped.

Two rules follow from that:

- A **tighter** (higher) bound prunes more, so the search is faster.
- A bound that is **too high** — above a score that really is reachable down that branch — makes BnB throw away the true best answer. Bounds must never do this. The word for "never too high" is *admissible*.

## Current state (after #64)

When a problem has unconfirmed ancestors, `LowestFee::bound` gives up its tight reasoning and returns a plain floor:

```
fee_floor = max( rate·(child weight) + ancestor_bump_lower_bound,
                 absolute fee,
                 RBF minimum )
```

The rest is switched off, because the tight path leans on two assumptions that ancestors break:

1. *"Selecting more never un-funds a selection."* False — a coin can drag in an unconfirmed parent whose bump costs more than the coin is worth.
2. *"A candidate costs its own weight."* False — a coin sharing an already-paid-for ancestor is cheaper than it looks; a coin dragging its own unpaid chain is dearer.

## Why it matters

Same candidates, ancestry declared vs not:

| problem | rounds |
| --- | --- |
| n=14, no ancestry (tight path) | 3304 |
| n=14, with ancestry | 6945 |
| n=16, no ancestry (tight path) | 8000 |
| n=16, with ancestry | 13769 |

Roughly **2× the work**. Answers are correct, just slower. (Not perfectly controlled — the optima differ because ancestors genuinely cost more — so read it as a rough ceiling on the headroom.)

## Plan

**1. Marginal cost helper.** Add "what would adding candidate `i` cost this selection *right now*?" = its own input weight priced at the target rate, plus the **extra** ancestor bump it brings given what is already dragged in. Cheap thanks to the private/shared split from #64: for privately reachable ancestors the delta is a fixed per-candidate pair; only shared ones need a lookup.

**2. Value-shortfall relaxation.** Rebuild the "resize" idea on marginal cost instead of raw value-per-weight: walk unselected candidates by best marginal value per marginal cost until funded, then allow a fractional slice of the last one. The ordering is only a heuristic once ancestors exist, so this step must **never** return `None` (see step 4).

**3. Funded-node bound.** Today a funded node's own score is not used, because a descendant can owe *less* (an ancestor that overpays nets against the deficit). Use `score − maximum reachable surplus` instead — the surplus figure already exists as `ancestor_bump_lower_bound`. Admissible, and much closer than the floor.

**4. Leave the `None` prunes off unless proven.** `select_iter().find(is_funded)?` returning `None` claims "nothing in this subtree can be funded". That is invalid when funding is not monotone, so it stays disabled.

## Watch out

#56 reports that the existing resize relaxation is **already inadmissible on master** (the greedy prefix's segwit corrections inflate `ideal_fee`). Re-enabling it for ancestors must not inherit that bug. Fixing #56 first, or building step 2 on marginal costs, both sidestep it.

## Done when

- The proptests in `tests/ancestor.rs` stay green: every bound ≤ every descendant score, `None` really means no solution in the subtree, and BnB == brute force. These catch exactly this class of mistake — three of them failed when the bound was deliberately made too tight.
- `run_bnb_lowest_fee_ancestors/{private,shared}/{20,50,100}` (benches added in #64) drop, closing most of the ~2× gap. Rough target: within ~1.2× of the no-ancestor path.

### LINKED ISSUE #66: Changeless<LowestFee> prunes away changeless solutions that max_weight forces
## What `Changeless` is meant to do

`Changeless<M>` finds the best selection that has **no change output**. It asks the inner metric (usually `LowestFee`) whether a selection would have change; if it would, that selection doesn't count.

To avoid searching pointlessly it has a shortcut, `change_unavoidable`: *if every selection reachable from here would have change, skip this whole branch.* It decides that by building the reachable selection with the **smallest excess** — add every remaining coin with negative effective value, since those shrink the excess. If even that one still has change, prune.

## The bug

There is a second route to being changeless that the shortcut does not know about.

`LowestFee` refuses to add change when the change output would push the transaction over `Target::max_weight`. So a selection can be changeless **purely because change would not fit**. The shortcut only reasons about excess, so it prunes branches containing exactly those solutions.

Net effect: BnB reports *no solution* while a perfectly good changeless solution exists.

## Verified reproduction (master, `abfb0ed`)

Three candidates, `max_weight = 710`, 1 sat/vb, target value 10 000, `DrainWeights::TR_KEYSPEND`:

```rust
let candidates = vec![
    Candidate { value: 121_861, weight: 200, input_count: 1, is_segwit: true },
    Candidate { value:   1_000, weight: 500, input_count: 1, is_segwit: true },
    Candidate { value: 193_757, weight: 368, input_count: 1, is_segwit: true },
];
let target = Target {
    fee: TargetFee { rate: FeeRate::from_sat_per_vb(1.0), absolute: 0, replace: None },
    outputs: TargetOutputs { value_sum: 10_000, weight_sum: 100, n_outputs: 1 },
    max_weight: Some(710),
};
let metric = || Changeless(LowestFee {
    long_term_feerate: FeeRate::from_sat_per_vb(1.0),
    dust_relay_feerate: FeeRate::from_sat_per_vb(1.0),
    drain_weights: DrainWeights::TR_KEYSPEND,
});

let mut a = CoinSelector::new(&candidates);
let expected = common::exhaustive_search(&mut a, target, &mut metric());

let mut b = CoinSelector::new(&candidates);
let found = common::bnb_search(&mut b, target, metric(), usize::MAX);

assert_eq!(expected.map(|(s, _)| s), found.ok().map(|(s, _)| s));
```

```
solution=[0✔, 2✔, 1☐], score=305618          <- brute force
exhaustive = Some(Ordf32(305618.0))
bnb        = Err(RoundLimit { max_rounds: 18446744073709551615, rounds: 2 })
```

`rounds: 2` matters: the queue **emptied**, so this is a prune, not a round limit. Unlimited-rounds BnB should be an exact detector here.

## Mechanism

| selection | weight, no change | weight, with change | change? | `Changeless` score |
| --- | --- | --- | --- | --- |
| `{0}` | 342 | 514 (fits) | yes, 111 732 | rejected — has change |
| `{0,2}` | 710 (exactly the cap) | 882 (**over** cap) | no | **305 618** |

At node `{0}`, `LowestFee` wants change, so `change_unavoidable` looks for a reachable selection with less excess. The only remaining coin (candidate 1) has *positive* effective value (+875), so nothing is added; the "least excess" selection is `{0}` itself, which still has change → **prune**. But `{0,2}` sits under `{0}`, and it is changeless because its change would weigh 882 > 710.

## Fix options — input wanted

1. **Make the shortcut cap-aware.** Don't prune when a reachable selection could exceed the cap *with* change weights. Cheap conservative test: if selecting everything still fits under cap-with-change, prune as today; otherwise don't. Keeps the prune wherever it is valid.
2. **Drop the shortcut whenever `target.max_weight.is_some()`.** Simplest and correct; loses pruning on every capped problem.
3. **Have the inner metric say *why* it refused change** (not worthwhile / dust / too heavy) so the wrapper can reason instead of guessing. More invasive, but removes the guesswork for good — and any future reason for refusing change won't silently reintroduce this.

My read: (1) is the best cost/benefit, (3) is the honest fix. Which do you prefer?

## Notes

- Unrelated to unconfirmed ancestors — reproduces on plain master, and predates #64.
- Distinct from #17 (`Changeless` usefulness when combined), though both are about this wrapper.
- In #64 the new `Changeless` proptest forces `max_weight: None` with a comment pointing here, so this stays visible rather than hidden.

evanlinjin:
## Summary

Makes coin selection pay for unconfirmed ancestors (CPFP) without putting ancestry on `Candidate`.

- **Problem model**: `SelectionProblem` owns the target, candidates, and ancestor graph; `CoinSelector` borrows it. `Input<Txid>` identifies the transaction each candidate input spends, and `AncestorToBump<Txid>` supplies ancestor weight, fee, and parent links.
- **Fee obligation**: selecting a coin charges the shortfall of the **union** of ancestors dragged in by the selection. Each ancestor is counted once, all ancestor weight and fee are netted together, and the bump saturates at zero.
- **Score**: the metric still scores the child transaction's fee. The bump is already inside the amount the child must pay, not added again afterward.
- **Private/shared split**: ancestors reachable through one candidate are folded into a per-candidate `(weight, fee)` pair. Only ancestors shared by several candidates need bitset de-duplication while scoring a selection.
- **BnB safety and speed**: exclusion batching requires matching `drags_in`; `SelectionView` maintains delta-aware per-branch aggregates so metric evaluation does not repeatedly scan selected candidates.
- **Tighter `LowestFee` bound**:
  - A funded node uses its score minus ancestor surplus still reachable by a descendant, clamped to the monotone fee floor.
  - An unfunded node independently relaxes the target-rate, absolute-fee, and RBF constraints into a minimum added **child input weight**, then evaluates the fee floor at that weight.
  - Candidate-specific ancestor cost is deliberately ignored in the resize: existing package surplus can absorb a later private deficit, so only `ancestor_bump_lower_bound` is valid for every descendant.
  - The ancestor path never returns `None` from funding heuristics. "Selecting everything is unfunded" does not prove that every subset is unfunded.
- **Dedicated changeless metric**: `LowestFeeChangeless` replaces the loose generic `Changeless<LowestFee>` composition. It reuses `LowestFee` funding and change decisions, then adds a monotone selected-value bound for pools up to 24 candidates. Larger pools retain `LowestFee` ordering to avoid finite-budget search starvation.

Ancestor weight does not count toward the child transaction's `max_weight`, and RBF rule 4 prices only child weight.

Closes #24.
Closes #65.
Closes #66.

## Dependencies

This self-contained branch includes the prerequisite work developed in:

- #59: bind a selector to one target
- #63: exact grouped legacy/segwit input weight pricing
- #53: delta-aware BnB metric evaluation

#60 is the earlier cluster-based exploration; this PR is the simpler no-cluster design.

## Out of scope

- Package feerate and TRUC package-size limits as first-class APIs
- #67: an apples-to-apples ancestor-aware benchmark against Bitcoin Core

## Benchmark findings

Full report: [coinselect-benchmark/FINDINGS.md](https://github.com/evanlinjin/coinselect-benchmark/blob/master/FINDINGS.md)

- Ancestor union bumps agree exactly with Bitcoin Core's post-discount combined bumps across all measured runs.
- coin-select reaches its oracle optimum on all eight brute-force-sized fixtures; Core misses four ancestry optima because overlap discounts are applied only after selection.
- The dedicated changeless bound makes `nested_ancestry_20` optimal in 200 rounds and `subsidizing_ancestry_20` optimal in 886 rounds, versus no result at 100,000 rounds previously.
- The large-pool ordering fallback preserves the pinned `subsidizing_ancestry_100` results: score `19006` at 100,000 rounds and `18925` at 20 million rounds.

## Test plan

- [x] Every commit independently passes `cargo fmt --all -- --check`
- [x] Every commit independently passes `cargo check --all-targets --all-features` and `cargo check --no-default-features`
- [x] Every commit independently passes `cargo clippy --all-targets --all-features --tests` with warnings denied
- [x] Every commit independently passes `cargo test --release` and `cargo doc --no-deps` with warnings denied
- [x] Every commit independently builds in release mode on Rust 1.54 with dev dependencies excluded, matching CI
- [x] 29 ancestor tests, including union, bump-floor, bound-admissibility, BnB-vs-exhaustive, and changeless-vs-exhaustive proptests
- [x] 10,000-case release stress runs for bound admissibility and BnB-vs-brute-force
- [x] Independent exhaustive oracle over 30,000 private, 30,000 shared, and 30,000 mixed-ancestry fixtures for both bound and BnB checks
- [x] Regressions for package surplus subsidizing a private deficit, absolute/RBF double counting, child-only RBF weight, and large-float cancellation
- [x] Criterion benchmarks for private/shared ancestry at 20, 50, and 100 candidates
- [x] Mutation-checked: ancestor union, bump lower bound, BnB look-alike batching, and changeless ancestor handling

## Review guide

1. `SelectionProblem::new`: parent walk and private/shared split.
2. `CoinSelector::ancestor_bump`: union accounting and package netting.
3. `CoinSelector::ancestor_bump_lower_bound`: why every descendant owes at least this amount.
4. `LowestFee::bound_with_ancestors`: funded surplus credit and the child-weight fractional relaxation.
5. `LowestFeeChangeless`: the monotone no-change bound and large-pool ordering fallback.
6. `tests/ancestor.rs` and `tests/lowest_fee_changeless.rs`: exhaustive admissibility and BnB-equivalence oracles.


evanlinjin:
Superseded by #85, which redoes this on top of the #59 → #63 → #84 stack. #84 replaced the best-first search this branch was built on, so the ancestor work is ported rather than rebased: the per-branch `SelectionView` cache becomes running totals on `CoinSelector`, and #75's sparse ancestor sets and #73's lookahead prune are included from the start.

PR #68, opened 2026-08-14
perf: search branch and bound with in-place DFS

evanlinjin:
## Summary

Replace best-first branch-and-bound (a `BinaryHeap` of cloned branches) with in-place DFS that visits the better-bound child first and backtracks.

- One selector/cache pair, mutated and undone, instead of cloning every node onto a heap.
- Equal bounds still prefer inclusion, same as the old tie-break.
- Exclusion still batches interchangeable candidates (same value, weight, and `drags_in`).
- Under a round cap this completes selections on large pools where the old frontier often spent the budget without a solution.

Public API is unchanged (`BnbIter` stays `pub(crate)`).

## Based on #64

This branch is stacked on #64 (`feature/ancestor-aware-selection-no-clustor`). Merge #64 first; the DFS change is the tip commit `perf: search branch and bound with in-place DFS`. Review that commit (or retarget this PR onto #64's branch) for a DFS-only diff.

## Benchmarks (`cargo bench -- run_bnb --quick`)

Versus best-first on the same #64 tree:

| bench | best-first | DFS |
|---|---|---|
| `run_bnb_lowest_fee` 20 / 50 / 100 | 85 µs / 2.4 ms / 6.0 ms | 43 µs / 1.2 ms / 4.4 ms |
| `run_bnb_lowest_fee_exhaust_cap` 200 / 500 / 1000 | 94 / 97 / 120 ms, **no solution** | 34 / 65 / 113 ms, **finds solutions** |
| ancestors private 100 | 130 ms | 54 ms |
| ancestors shared 100 | 156 ms | 54 ms |

DFS also finds a `LowestFee` solution on a 10k-coin pool within the 100k-round cap.

Some instances visit more nodes (exact-match regression: 3,194 → 62,452 rounds) but wall-clock still improves from skipping heap/clones.

## Test plan

- [x] `cargo fmt -- --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test --release` (bnb, lowest_fee, lowest_fee_changeless, ancestor, including exhaustive/proptest oracles)
- [x] Criterion `run_bnb*` benches vs the previous best-first implementation


evanlinjin:
This is AI-generated slop that actually produces worse results.

PR #70, opened 2026-08-14
feat: seed branch and bound with a greedy incumbent

evanlinjin:
> [!IMPORTANT]
> **Stacked on #69, which is itself stacked on #64.** This branch is one commit on top of #69's head (`fb5a021`), and this repo doesn't carry either branch, so the diff below currently includes both. Review after they merge, at which point this reduces to the single commit described here. Draft until then.

Branch and bound is not anytime. A caller whose round budget runs out before the first complete selection gets `NoBnbSolution::RoundLimit` and falls through to whatever fallback it has — on a large pool that is far worse than the selection a single greedy pass would have handed it for free. In the benchmark matrix this is not a corner case: on the 500–2000 candidate fixtures the search regularly spends its entire 100,000-node budget on shallow branches and returns nothing.

This yields the greedy selection before expanding the first node, and adopts its score as the incumbent.

```rust
fn seed_greedy_incumbent(&mut self, selector: &CoinSelector<'a>) {
    let mut seed = selector.clone();
    if seed.select_until_target_met().is_err() {
        return;
    }
    if let Some(score) = self.metric.score(&seed.compute_view()) {
        self.best = Some(score);
        self.seed = Some((seed, score));
    }
}
```

Only the incumbent changes, never the bound, so the optimum stays reachable and the improving-solutions contract is unaffected — the seed is simply the first solution the iterator yields, and everything after it still has to beat it.

It yields nothing for a metric that rejects the greedy prefix outright. `LowestFeeChangeless` is the case: overshooting the target is exactly what a greedy pass does and exactly what that metric will not score, so `RoundLimit` still means for it what it meant before.

## Measured

[`coinselect-benchmark`](https://github.com/evanlinjin/coinselect-benchmark) `bench.py compare-revs`, 126 fixture/track pairs, 100,000-node budget, measured on this branch against #64's head:

| | value |
|---|---|
| solutions **gained** | **13** |
| solutions lost | **0** |
| selections changed | **0** |
| scores changed | **0** |
| node counts | flat (1.00x) |
| peak RSS / wall clock | flat |

The 13 are the searches that previously exhausted their budget and returned nothing:

```
no_ancestry_500 / _1000 / _2000          changeful
shared_ancestry_200 / _500 / _1000 / _2000   changeful
wallet_mixed_500 / _1000 / _2000         changeful, wallet
```

Not one selection anywhere else in the matrix changes, which is what you would expect from a change that only supplies an incumbent: where the search already finished, it finished at the same answer.

The two branches compose exactly — measured together, the stack is #69's node reduction plus these 13 recovered solutions, with no interference. One changes the bound, the other the incumbent.

## Notes

- One behavioural change worth calling out in review: `NoBnbSolution::RoundLimit` is now much rarer. It means the metric rejected the greedy selection too, rather than "we ran out of rounds".
- The two round-count assertions in `tests/bnb.rs` each move by one — the seed is a round.
- New test `run_bnb_returns_the_greedy_selection_on_a_tight_budget`: 500 candidates, `max_rounds = 1`, must still come back funded.
- Passes `cargo fmt`/`check`/`clippy`/`doc`/`test` and `--no-default-features`.

https://claude.ai/code/session_019qxj8PzFNeqeKw7XkuzvK3


evanlinjin:
Superseded by the new stack. The greedy incumbent seed is in #84 (c495ffc), and the lookahead prune this branch also carried is in #85 (5a3cbb0). The changeless window cut on top is not ported: #84 removed the changeless metrics, so it has no caller — along with `ancestor_bump_upper_bound`, which that cut was the only consumer of.

PR #73, opened 2026-08-15
feat!: depth-first branch and bound, and remove the changeless metrics

evanlinjin:
Draft. Two things, on top of the ancestor-aware selection stack: search branch and bound **depth-first**, and **remove the changeless metrics**.

Benchmarked with [coinselect-benchmark](https://github.com/evanlinjin/coinselect-benchmark), which scores every selection from the fixture with one fee model rather than trusting either engine's own numbers.

## Depth-first search

The priority-queue traversal has a structural problem: both shipped metrics' bounds increase with depth, so a min-heap always pops the shallowest node. That is uniform-cost search with a branching factor equal to the candidate count — it expands every 1-input prefix, then every 2-input prefix, and on a large pool the budget is gone before it reaches any funded leaf. Solutions in the benchmark matrix are 1–28 inputs deep; they are not hard to reach, they are a dozen levels down a tree the search never descends.

Depth-first with in-place backtracking reaches a funded leaf in as many expansions as the solution has inputs, and carries one path instead of a frontier.

Measured on the wallet track, equal **wall-clock** budgets with the round cap lifted, package fee from the shared model, 42 fixtures:

| 1000 ms budget | total package fee | vs best-first | peak RSS |
| --- | --- | --- | --- |
| best-first | 1,441,766 | — | 555.7 MB |
| **depth-first** | **1,379,617** | **−4.31%** | **3.6 MB** |

Cheaper selections *and* a factor of 150 less memory. The memory gap widens with the budget — best-first goes 7 MB → 58 MB → 554 MB as the budget grows 10× twice, while depth-first stays flat, because a best-first search over a depth-increasing bound never finishes a level and so never discards one. Given a 130-second budget the old frontier reached 20 GB on a 500-candidate problem; the round cap was doing double duty as a memory guard.

## Removing the changeless metrics

`LowestFee` already decides for itself whether a selection should carry change — adding one only when it lowers the long-term fee, clears the dust threshold, and fits `max_weight`. A separate changeless objective duplicates that decision and then constrains it.

It also does not seed. `BnbIter` seeds a greedy incumbent so a search that runs out of rounds returns something rather than nothing; a greedy prefix taken in descending value-per-weight overshoots the target, so `LowestFee` wants change and the changeless metric rejects it — at every prefix length. Measured across the benchmark's 42 fixtures, the greedy seed is accepted by `LowestFee` **42 times** and by `LowestFeeChangeless` **0 times**. Nor is it a matter of picking a better prefix: the ascending prefix overshoots by the least of any greedy ordering and its smallest excess across the matrix is 1270 sat against a changeless ceiling of 580 sat — still 0 of 42.

This is a breaking change. Callers that required a changeless transaction should use `LowestFee` and inspect the returned `Drain`.

## What is not here

An `ancestor_bump_upper_bound` — a ceiling mirroring `ancestor_bump_lower_bound` — was implemented and measured on both traversals. It is left out on purpose: its only call site was the changeless window cut, so it becomes dead code with the metrics gone. For the record it was a byte-identical no-op on the wallet track (42/42) and its entire measured value was two kernel-track conversions under depth-first. It is one cherry-pick away if a changeless-style objective ever returns.

## A pre-existing optimality gap, characterised but not fixed

While benchmarking I found that best-first and depth-first can both report the tree exhausted and still disagree, on `no_ancestry_1000`. It is deterministic and reproducible at a 20-second budget with the round cap lifted, and it goes **both ways**:

| metric | best-first | depth-first | who missed the optimum |
| --- | --- | --- | --- |
| `LowestFee` (44 inputs, changeless, both exhausted) | 30463 | **30462** | **best-first**, by 1 sat |
| `LowestFeeChangeless` (44 inputs, both exhausted) | **30460** | 30462 | depth-first, by 2 sats |

Two exhaustive searches over one objective must agree, so on each row one of them terminated before finding the better selection. Note the direction on the row that matters: on `LowestFee` — the metric that survives this PR — **the shipped best-first search is the one that misses it, and depth-first is correct**. So this is a pre-existing gap that depth-first improves here, not something the traversal change introduces.

What I can rule out: it is not size alone (no fixture at n ≤ 20 shows it, and both searches match a brute-force oracle over all 2^n subsets on every one of them), and it is not ancestry (this fixture has no unconfirmed parents at all). The magnitudes — 1 and 2 sats against a ~30,000 sat score — point at `f32` arithmetic in the bound making it very slightly inadmissible, which would let the "best remaining lower bound exceeds the incumbent" termination fire one step early. I have not isolated that, so treat it as a hypothesis.

It is worth a reviewer's eye on the bound's precision rather than on the traversal.


evanlinjin:
Superseded by #84, which rebases the depth-first search and the changeless removal onto #63 so they land ahead of the ancestor-aware stack.

PR #75, opened 2026-08-16
perf!: store per-candidate ancestor sets sparsely

evanlinjin:
> **Draft, and stacked on #73.** One commit on top of it. Independent of #74 — both branch from #73 and they touch different code.

`SelectionProblem` stores `drags_in` and `shared_drags_in` as one dense `Bitset` per candidate over **every** ancestor, so the arrays cost candidates x ancestors bits each.

A candidate drags in its residing transactions and their unconfirmed parents. That is a handful of entries no matter how large the pool gets — measured across generated pools, **mean 0.42 to 0.60 entries and never more than two**. On a 200,000-candidate pool with 26,666 ancestors the dense array is 667 MB holding sets that are **0.002% full**.

## The cost is not only memory

Iterating a dense bitset is O(ancestors) per candidate however few bits are set. Building the selection cache walks every candidate's shared set, so construction is O(candidates x ancestors) in **time** as well as space — 5.3 billion bit positions where the entries number 84,000.

That is enough to lose a wall-clock budget outright. On `wallet_mixed_200000` the benchmark harness reported **no solution at a 100 ms budget, with zero nodes expanded**, and it took a while to work out why: setting up the search takes 464 ms, so the deadline expired before the first node. The search, the bound and the single-random-draw fallback were all fine and none of them ran. The greedy seed alone would have answered, if anything had polled the iterator.

## The change

Store them flat: one `Vec<u32>` of ancestor indices with a `Vec<u32>` of per-candidate offsets.

Every read of these sets is a full walk of one candidate's entries — `bump_of`, the cache's add and sub, the four `shared_drags_in` walks in `selection_view.rs` — plus one set-equality comparison in the exclusion plan. Nothing needs O(1) membership and nothing mutates them after construction, so a slice is all they have to be. Construction reuses a single scratch bitset rather than allocating one per candidate, so the cost being removed does not reappear while building.

`Bitset` is untouched where it is used over *candidates*: the selected and banned sets are dense and membership-tested constantly.

## Measured

| 200,000 candidates, 26,666 ancestors | peak RSS | search setup |
| --- | --- | --- |
| dense `Bitset` per candidate | 1,332 MB | 464 ms |
| flat indices + offsets | **58 MB** (-95.7%) | **54 ms** (8.5x) |

At 20,000 candidates, 21.6 MB down to 7.6 MB. On a 200,000-candidate pool with no ancestors at all nothing changes, which is the control behaving.

`wallet_mixed_200000` at a 100 ms budget goes from **no solution** to 9,217 inputs and 5,632 nodes.

## Correctness

- **Byte-identical on all 42 benchmark fixtures** — same selections, scores, round counts and exhausted flags.
- 80 tests green on `--all-features` and `--no-default-features`.

## Breaking

`drags_in` and `shared_drags_in` return `&[u32]` rather than `&Bitset`. Callers iterating them need `for &i in ...` and a `usize` cast; a set-equality check still works as a slice compare. In the benchmark harness this was a three-line change.

The alternative considered was a `BTreeSet` per candidate. It fixes the asymptotics, but for sets of nought to two elements it pays an allocation per non-empty set — roughly 100,000 of them here — and a pointer chase per read, where the flat layout is contiguous and allocates nothing per candidate.

https://claude.ai/code/session_01HLiTkESMktypGJhFag2ZBM


evanlinjin:
Included in #85 from the start: per-candidate ancestor sets are stored as sorted `&[u32]` slices out of one flat array there, so the dense-bitset cost this PR fixes never lands.

